PR #23, opened 2015-11-19
Adding comparison functions.

tubedude:
@ericmj I've created the `gt`, `lt`, `eq` functions but I was confused with your `cmp(num1, num2)` spec.
Do you suggest returning atoms as result of the comparison?

### DIFF
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -290,6 +290,20 @@ defmodule Decimal do
   end
 
   @doc """
+  Compares two numbers numerically. If the first number is greater than the second
+  `:gt` is returned, if less than `:lt` is returned, if both numbers are equal
+  `:eq` is returned. Otherwise, if any number is a `NaN`, NaN is returned.
+  """
+  @spec cmp(t, t) :: :lt | :eq | :gt | :qNaN
+  def cmp(num1, num2) do
+    case compare(num1, num2) do
+      %Decimal{coef: 1, sign: -1} -> :lt
+      %Decimal{coef: 0} -> :eq
+      %Decimal{coef: 1, sign: 1} -> :gt
+    end
+  end
+
+  @doc """
   Compares two numbers numerically and returns `true` if they are equal,
   otherwise `false`.
   """
--- a/test/decimal_test.exs
+++ b/test/decimal_test.exs
@@ -188,6 +188,20 @@ defmodule DecimalTest do
     end
   end
 
+  test "cmp" do
+    assert Decimal.cmp(~d"420", ~d"42e1") == :eq
+    assert Decimal.cmp(~d"1", ~d"0")      == :gt
+    assert Decimal.cmp(~d"0", ~d"1")      == :lt
+    assert Decimal.cmp(~d"0", ~d"-0")     == :eq
+
+    assert_raise Error, fn ->
+      Decimal.compare(~d"snan", ~d"0")
+    end
+    assert_raise CaseClauseError, fn ->
+      Decimal.cmp(~d"nan", ~d"1")
+    end
+  end
+
   test "equal?" do
     assert Decimal.equal?(~d"420", ~d"42e1")
     refute Decimal.equal?(~d"1", ~d"0")


lib/decimal.ex
```diff
@@ -2,8 +2,8 @@ defmodule Decimal do
   @moduledoc """
   Decimal arithmetic on arbitrary precision floating-point numbers.
 
-  A number is represented by a signed coefficient and exponent such that: `sign
-  * coefficient * 10^exponent`. All numbers are represented and calculated
+  A number is represented by a signed coefficient and exponent such that: ```sign
+  * coefficient * 10^exponent```. All numbers are represented and calculated
```
ericmj: Please keep as-is, the previous markdown was correct.

ericmj:
I prefer a function returning the `:eq | :lt | :gt` atoms. Maybe we can call it `compare`, it could be confusing with the current `cmp` though? /cc @josevalim

lib/decimal.ex
tubedude: @ericmj I was looking at this and thought it would fix it.
decimal_v1_1_0_documentation

ericmj: Looks like a bug in the markdown parser. Please confirm if you can reproduce it in master of https://github.com/pragdave/earmark/, if so please report the bug.

tubedude: I'll remove the edit for now until I'm able to figure this out:
```bash
$ mix docs
Compiled lib/decimal.ex
Generated decimal app
** (UndefinedFunctionError) undefined function: Error.exception/1 (module Error is not available)
    Error.exception([message: "module Decimal.Context was not compiled with flag --docs"])
    lib/ex_doc/retriever.ex:82: ExDoc.Retriever.verify_module/1
    lib/ex_doc/retriever.ex:69: ExDoc.Retriever.get_module/2
    (elixir) lib/enum.ex:1043: anonymous fn/3 in Enum.map/2
    (elixir) lib/enum.ex:1387: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/enum.ex:1043: Enum.map/2
    lib/ex_doc/retriever.ex:48: ExDoc.Retriever.docs_from_modules/2
    lib/ex_doc.ex:17: ExDoc.generate_docs/3
```

tubedude:
It think it is the other way around. The current function is called `compare/2` https://github.com/ericmj/decimal/blob/master/lib/decimal.ex#L277-L290
So the new one should be `cmp/2`.

lib/decimal.ex
ericmj: Try updating ex_doc and run on latest stable elixir.

josevalim:
:+1: for returning atoms. The current `compare` function is really weird though.

ericmj:
@josevalim The current compare follows the specification, should we drop it in favor of the atoms version?

lib/decimal.ex
tubedude: @ericmj It is probably a bug in the parser. I'll open a bug report there.
decimal_ _decimal_v1_1_1-dev
I was able to update ex_doc and it rendered the page with the same error.

josevalim:
@ericmj yes, please. :)

josevalim:
But I don't think we can just drop `compare`, can we? I would add `cmp` and deprecate `compare`. :)

ericmj:
I would rather deprecate it for some period and then replace it with this version. Alternatively just replace it and release 2.0.0.

tubedude:
So the PR is ready as it is.

josevalim:
@ericmj I would introduce `cmp`, deprecate `compare`. We can add `compare` in the future back or when we merge it into Elixir. Not worth releasing 2.0 because of this.

josevalim:
I agree then we should merge this as is anyway, we can deprecate in another PR. :)

ericmj:
@tubedude Please remove the eq, gt, and lt functions. Also don't pattern match on qNaN, just let it raise instead. After those two changes it is ready for merge. Thanks!

lau:
+1 for this. Is anything missing for this to be merged?

ericmj:
It's ready. I never got a notification about it being updated so I forgot about it, sorry about that.

lau:
No worries. I have experienced the exact same thing - waiting for an update to a PR, and then not realizing the update happens, because GitHub doesn't send notifications for that.

PR #78, opened 2018-03-14
Soft deprecate passing float to new/1 in favor of from_float/1

wojtekmach:
Ref https://elixirforum.com/t/gringotts-a-complete-payment-library-for-elixir-and-phoenix-framework/11054/40 cc @kipcole9 

I guess I should have started discussion first but it was so easy to create a patch (and thus determine impact) so here we are :) If we would be to ever hard deprecate it, a notable call to this is in Ecto: [Ecto.Type.cast](https://github.com/elixir-ecto/ecto/blob/v2.2/lib/ecto/type.ex#L322:L324) so it would warn a lot.

As part of the PR I added a commit that adds hard deprecations: https://github.com/ericmj/decimal/commit/8a22acc003d6eaf050289c22fb956dcde711b044 (reverted for now)

### DIFF
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -215,20 +215,6 @@ defmodule Decimal do
     end
   end
 
-  if Version.compare(System.version(), "1.3.0") == :lt do
-    defmacrop warn(message) do
-      quote do
-        IO.puts(:stderr, "warning: " <> unquote(message))
-      end
-    end
-  else
-    defmacrop warn(message) do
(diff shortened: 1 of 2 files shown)


wojtekmach:
@josevalim it should be straightforward to start transition for this in Ecto and I'll be happy to work on that when new decimal release is out - please let me know how does this sounds.

josevalim:
@wojtekmach sounds good to me!

PR #85, opened 2018-05-22
Mix: missing `float` in typespec

harlantwood:
This was caught by dialyzer in our project, which uses `decimal`.


wojtekmach:
Thanks for the report! We are soft deprecating using floats in Decimal.new and getting dialyzer warning is an unfortunate consequence, see https://github.com/ericmj/decimal/issues/81 for more info.

PR #109, opened 2018-12-10
fix(decimal): Add from float for integers

raksonibs:
I was using `Decimal.from_float/1` as `Decimal.new/1` in my code, but noticed its handling for integers was a bit off,  wherein Decimal.new would just accept a float or integer.

I thought it might be wise for from_float then be able to also handle integers, despite the name.

### DIFF
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -891,7 +891,7 @@ defmodule Decimal do
   @doc """
   Check if given number is positive
   """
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you for sending the PR, but I think accepting integers in a `from_float` function would be too confusing. If you're OK with handling both integers & floats in a single function I'd suggest to create a custom function that does exactly that.

PR #118, opened 2019-05-26
Update tests and docs to reflect from_any/1 accepting decimal input

eqmvii:
Small update to documentation, typspec, and tests requested in the discussion here: https://github.com/ericmj/decimal/pull/116#pullrequestreview-241607628

Since `new/1` can accept a decimal input, and is called by `from_any/1`, the tests and documentation now reflect that fact. 

Also updated its typespec to use the `decimal` type to account for any of decimal / integer / string.

### DIFF
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -1238,7 +1238,7 @@ defmodule Decimal do
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


wojtekmach:
Thank you!

PR #139, opened 2019-12-30
Soft-deprecate Decimal.parse/1

wojtekmach:


ericmj:
I think we should hard deprecate instead.

PR #144, opened 2020-01-07
Change Decimal.cast/1 to return {:ok, t} | :error

wojtekmach:


test/decimal_test.exs
```diff
@@ -119,6 +119,20 @@ defmodule DecimalTest do
...
+    assert Decimal.cast(123) == {:ok, d(1, 123, 0)}
+    assert Decimal.cast(123.0) == {:ok, d(1, 1230, -1)}
+    assert Decimal.cast("123") == {:ok, d(1, 123, 0)}
+    assert Decimal.cast(d(1, 123, 0)) == {:ok, d(1, 123, 0)}
+
+    assert Decimal.cast("one two three") == :error
+    assert Decimal.cast("e0") == :error
+
+    assert_raise FunctionClauseError, fn ->
+      Decimal.cast(:one_two_three)
```
wojtekmach: we could also return `:error` here instead. I'm fine with either approach.

ericmj: If we do return `:error` we should change the spec to accept any input type.

wojtekmach:
Right. I’m still on the fence so it’s up to you how we should handle it :)

ericmj:
We could make it work like ecto which accepts any input type and returns `:error`. `parse/1` would also return `:error` and `new/1` would raise.

wojtekmach:
Sounds good.

PR #172, opened 2021-07-09
Add `with_decimal/1` macro

yunmikun2:
Here we add convinience macro that makes it easier to write arithmetic
expressions with `Decimal`.

The macro replaces arithmetic operators (`+`, `-`, `*`, and `/`) with
respective decimal functions (`add`, `sub`, `mult`, and `div`). It
takes advantage of the fact that all these functions can convert
integers to `%Decimal{}` struct authomatically so in the most common
case you don't need to convert the values manually. But in case you do
need (e.g. when you convert float value) you can use functions
available in `Decimal` module. The same applies to specific functions
like `Decimal.div_int/2` and `Decimal.round/3`.

Example:

    percent = 20

    value =
      with_decimal do
        Decimal.from_float(100.0) * percent / 100
      end

    value # #Decimal<20.0>

### DIFF
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -1770,6 +1770,50 @@ defmodule Decimal do
   else
     defp integer_to_charlist(string), do: Integer.to_charlist(string)
   end
+
+  @doc """
+  Replaces all `*`, `/`, `+`, and `-` with respective `Decimal` functions.
+
+  ## Examples
+
+      iex> import Decimal, only: [with_decimal: 1]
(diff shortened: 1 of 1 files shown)


yunmikun2:
@ericmj, oh, sorry. I should've looked at existing closed issues first. :sweat_smile:

wojtekmach:
No worries, thanks for the PR! If you'd like to explore a separate package, one of the earlier issues is https://github.com/ericmj/decimal/issues/145 and it has some valuable context that might be useful.

Issue #175, opened 2021-08-31
Proposal: Change the Inspect protocol implementation


Adzz:
When I a Decimal outputted to the console it looks like this:

```elixir
#Decimal<1>
```

This is fine and all but makes copy / pasting from the terminal a nightmare as the above isn't valid elixir code. Other structs handle this by having a sigil - Date for example. That means you can output `"~D[2020-01-01]"`

Would it be possible to change the inspect protocol to instead output this:

```elixir
Decimal.new("1")
```

Or have some other way to make this better. I know the idea of a Decimal sigil has been rejected before -is there any movement on that? I know Decimal even uses a [sigil in its tests ](https://github.com/ericmj/decimal/blob/master/test/test_helper.exs)

Adzz:
Nice! I'd be happy to try it for sure.

Adzz:
https://github.com/ericmj/decimal/pull/176

Adzz:
Hey 👋  Are there any plans to do a release for this?

ppraisethesun:
Elixir 1.14 is moving towards expression based inspects, so it would be a good time to release this

PR #176, opened 2021-08-31
Improve the Inspect protocol implementation

Adzz:
Currently the inspect protocol is implemented such that it prints
Decimals that look like this:

```sh
#Decimal<2.1>
```

This is great, but can be tricky when copy / pasting some printed output
into a repl for example, because `#Decimal<2.1>` is not valid elixir
syntax.

Other structs handle this by having a sigil - here we sidestep the
discussion of "should Decimal have a sigil" by implementing the Inspect
protocol such that it returns valid elixir code:

```elixir
Decimal.new("2.1")
```

Not sure if this should be considered a breaking change.

Closes https://github.com/ericmj/decimal/issues/175

### DIFF
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -203,10 +203,10 @@ defmodule Decimal do
   ## Examples
 
       iex> Decimal.add(1, "1.1")
-      #Decimal<2.1>
+      Decimal.new("2.1")
 
       iex> Decimal.add(1, "Inf")
-      #Decimal<Infinity>
+      Decimal.new("Infinity")
 
   """
   @spec add(decimal, decimal) :: t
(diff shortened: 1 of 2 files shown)


ericmj (approved):

wojtekmach:
Looks good to me. I plan to re-send my multi-character sigil proposal soon, so that we could instead do e.g. `~Decimal<1>`, so let's hold on this PR for a while. If the proposal is not accepted, let's definitely move forward with this PR.  Thanks!

wojtekmach:
@Adzz oops, sorry for the radio silence. I'll try to send the proposal soon, I think I can add important context from the previous attempts at it. Thanks for the offer!

wojtekmach:
In https://groups.google.com/g/elixir-lang-core/c/RA5uXkilYsE there was an idea for "constructor" syntax: `%Mod[item1, item2, ...]` syntax. For Decimal this would mean: `%Decimal["1.0"]`. If we have that then it doesn't make sense to have `~Decimal[1.0]` too, so I won't be submitting the proposal for that after all. There are many challenges with implementing `%Mod[item1, ...]` so let's do [what Elixir does](https://github.com/elixir-lang/elixir/issues/11610) and that is to print the expression representation in the meantime. Thanks for the PR!

PR #186, opened 2022-08-31
Add zero and sum functions

urcadox:
Hello,

I have had to sum a list of decimals a few times and find it a bit tedious to always have to write the repetitive albeit simple reduce code.

Having to use `Decimal.new(0)`as the default value of the reducer also feels wrong somewhat.

This PR fixes both of these issues. It adds a `Decimal.zero/1` function which gives you the default zero value of the Decimal type, and a `Decimal.sum/1` function which takes an enumerable of Decimal and returns their sum.

I'm not an experienced Elixir developer, so this may not be the best way to do this in Elixir. I would have loved to find a way to implement `Enum.sum` for Decimal, but that would require overloading the `+` operator which does not seem like a good idea.

Thanks for your feedback!

### DIFF
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -1160,6 +1160,18 @@ defmodule Decimal do
   end
 
   @doc """
+  Returns the default zero value of a Decimal.
+
+  ## Examples
+
+      iex> Decimal.zero()
+      Decimal.new(0)
+
+  """
+  @spec zero() :: t
+  def zero(), do: Decimal.new(0)
+
+  @doc """
   Creates a new decimal number from the sign, coefficient and exponent such that
   the number will be: `sign * coefficient * 10 ^ exponent`.
 
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -1015,6 +1015,20 @@ defmodule Decimal do
   end
 
   @doc """
+  Returns the sum of all decimals in the enumerable.
+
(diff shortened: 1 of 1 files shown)


lib/decimal.ex
```diff
@@ -1014,6 +1014,20 @@ defmodule Decimal do
...
+
+  ## Examples
+
+      iex> Decimal.sum([Decimal.new("1.234"), Decimal.new(2)])
+      Decimal.new("3.234")
+
+  """
+  @spec sum([decimal]) :: t
+  def sum(decimals) do
+    Enum.reduce(decimals, Decimal.zero(), fn d, acc -> Decimal.add(d, acc) end)
```
J3RN, line 1028: For whatever it's worth, if you don't mind the shorthand lambda syntax, you could condense `sum` to this:
```suggestion
    Enum.reduce(decimals, Decimal.zero(), &Decimal.add/2)
```

J3RN (commented):
I came here to make the same PR! :100:

We actually introduced an `Amount.zero/0` function into our codebase to serve the purpose of your `Decimal.zero/0`.

lib/decimal.ex
urcadox: Hey @J3RN,

I don't mind the lambda syntax at all, I'm just not used to see it for an arity > 1, this is completely fine though.

Issue #213, opened 2024-11-14
Exception returned when `Decimal.parse("7654832e-d977-469d-a88d-f27d738d4097")`


abel-cherinet-clark:
Here is what the error looks like when running `Decimal.parse("7654832e-d977-469d-a88d-f27d738d4097")`

```
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not a textual representation of an integer

    (erts 14.0.2) :erlang.list_to_integer(~c"-")
    (decimal 2.1.1) lib/decimal.ex:1873: Decimal.parse_unsign/1
    iex:1: (file)
```

PR #218, opened 2025-01-16
Fix parsing of 1e-d

treere:
Hi,
in this PR i fix https://github.com/ericmj/decimal/issues/213 .

I am returning value and not an `:error` because the first part is a valid number

Thank you! :)


ericmj (approved):

wojtekmach:
Thank you!

PR #227, opened 2026-05-07
 Pass opts through to parse from Decimal.new/2

bkilshaw:
The 3.0 release introduced passing opts to Decimal.parse but Decimal.new/1 didn't expose it, so callers had no way to use those options when constructing a Decimal from a string via new.

This adds an optional opts argument to Decimal.new, forwarding it to parse/2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


ericmj:
Thanks for the PR! Can you add a test for this also?

PR #235, opened 2026-05-26
Fix Elixir 1.20 compilation warnings

gilbertwong96:
Resolve compilation warnings introduced in Elixir 1.20.

- Removed unreachable `check_dbl_min_max(%Decimal{coef: :inf})` clause. The only call site in `to_float/1` guards on `is_integer(coef)`, so the `:inf` pattern can never match.


ericmj:
Thank you!

PR #240, opened 2026-08-14
Performance

tomciopp:
Rewrites the hot paths of the library: division, context rounding, `round/3`, comparison, and power-of-ten computation to use native bignum arithmetic instead of digit at a time loops and charlist manipulation. No public API changes, no behavior changes; results, struct representations, and signal flags are byte-identical to the current implementation, verified by the existing suite plus new unit and property tests added in this PR.

**Headline numbers** (M1, Elixir 1.19.5 / OTP 29, benchmarked with the extended `bench.exs` in this PR):

| Operation | Before | After | Speedup |
|---|---|---|---|
| `div` (mixed sizes) | 230.6 ms | 65.6 ms | **3.5x** |
| `div` (34-digit operands) | 3.93 ms | 0.86 ms | **4.6x** |
| `add` (34-digit) | 3.80 ms | 1.20 ms | 3.2x |
| `mult` (34-digit) | 1.31 ms | 0.59 ms | 2.2x |
| `add` / `sub` | ~115 ms | ~58 ms | 2.0x |
| `mult` | 65.9 ms | 34.8 ms | 1.9x |
| `round` | 4.91 ms | 2.28 ms | 2.2x |
| `compare` | 2725 ms | 1502 ms | 1.8x |
| `sqrt` | 7.00 ms | 5.07 ms | 1.4x |
| `to_string` | 1.43/1.64 ms | 1.13/1.23 ms | ~1.2x |

Memory allocation drops are larger than the speedups: `compare` −94% (1135 MB → 68 MB per benchmark iteration), `div` −82%, `add` −83%, `mult` −77%, `sqrt` −89%.

---

### 1. `bench.exs`: cover the full operation surface

The benchmark previously measured only `compare/2`. It now has 14 jobs: `add`, `sub`, `mult`, `div` over a deterministic sample of the operand matrix; the same four over 34-digit (decimal128-precision) operands where division and rounding costs dominate; `round`, `normalize`, `sqrt`, `to_string` (`:scientific` and `:normal`); and a same-scale `compare` workload (equal exponents, equal coefficient lengths — the shape of money comparisons) that always reaches coefficient alignment instead of short-circuiting on the adjusted exponent. The original `compare` job keeps its exact workload so previously saved results remain comparable.

### 2. `compare/2`: no padding when exponents are equal

`pad_num/2` multiplied **both** coefficients by at least 10 on every coefficient-level comparison (`pow10(max(n, 0) + 1)`), even when the exponents were already equal. The comparison branch now reuses `add_align/4`: equal exponents compare coefficients directly with zero multiplications; unequal exponents multiply only the larger-exponent side by `pow10(diff)`. Since this branch is only reached when adjusted exponents are equal, the exponent gap is bounded by the coefficient-length gap, so the padding stays small. `pad_num/2` had no other callers and is deleted.

### 3. `pow10/1`: binary powering; `decimal_power10/1` deleted

Above its 104-entry compile-time table, `pow10/1` recursed as `pow10(104) * pow10(n - 104)` — a linear chain multiplying an ever-growing bignum, quadratic overall. It now uses square-and-multiply (recurse on `n >>> 1`, square, multiply by 10 when odd): O(log n) multiplications.

This also let us delete `decimal_power10/1`, which computed 10^n by **building and parsing a string** (`binary_to_integer("1" <> zeroes(n))`). That function sat in `integer_decimal_digit_count/2`'s verification step — meaning every digit count of a >18-digit coefficient paid one or two string-build-and-parse cycles. This predates the PR and was already the dominant allocation cost in `compare` (via `adjust_exp`); an intermediate benchmark run during development showed it regressing add/mult by 36–80% once the new rounding fast path leaned on digit counting harder, which is how it was found. All three call sites now use `pow10/1` (a table lookup for anything ≤104 digits). This single change accounts for most of the memory reduction across the library.

### 4. Division: one native division instead of a digit loop

`div_calc/5` computed each quotient digit by repeated subtraction — up to 9 interpreted bignum subtractions plus two multiplications per digit, ~300 bignum operations per division at precision 34. The loop is equivalent to a single scaled division, so it is now one:

- `div_adjust/2` aligns the operands to `coef2 <= coef1 < 10 * coef2` by computing the shift from `coef_length/1` digit counts (O(1)) instead of one power of ten per iteration.
- `div_calc/4` computes `div(coef1 * pow10(precision), coef2)`: given the alignment invariant, the quotient has **exactly** `precision + 1` digits (the last being the guard digit), and the remainder is the sticky bit carried into rounding, exactly as before.
- Exact quotients strip the trailing zeros the scaling introduced — but never below a non-negative final adjust — reproducing the loop's preferred-exponent behavior (`div(100, 1)` is still `100`, not `1E+2`).
- `div_int_calc/5` is deleted. `integer_division/5` uses the same closed form, and because the quotient's digit count is known up front (`exp1 - exp2 - adjust + 1`), quotients over the precision limit are rejected **before** being materialized. The old loop's iteration count was proportional to the exponent gap, unbounded for hostile input; the new check is O(1).

Two loop quirks are deliberately preserved (and now pinned by tests):

1. An exact quotient whose adjust cannot reach zero (dividend vastly larger than divisor, e.g. `div(1e40-as-coefficient, 1)`) signals `[:inexact, :rounded]` exactly as the loop did — arguably a latent bug, kept for parity and documented in a test so any future fix is a conscious decision.
2. The `base10?` signal check runs on the closed-form remainder, which in some alignments is 10x the loop's remainder — equivalent, since being a power of ten is invariant under that scaling, and the remainder is otherwise only used as a boolean sticky bit.

### 5. Context rounding (`precision/4`): O(1) fast path, integer arithmetic

This function runs on virtually every arithmetic result. It previously started with `:erlang.integer_to_list(coef)` + `length/1` — an O(digits) conversion and traversal — before discovering, in the common case, that nothing needed rounding. It now:

- gets the digit count from `coef_length/1` (a guard-clause chain up to 18 digits, bit-length estimate beyond), and returns the number untouched with **zero allocation** when it fits the precision and no sticky bit is set;
- when rounding is needed, extracts significand / guard digit / rest-nonzero with `div`/`rem` via the new `split_digits/3` helper instead of `:lists.split`;
- increments by `signif + 1` instead of walking a reversed charlist carrying nines;
- detects the carry-lengthening case (from #238) as `signif == pow10(precision)`.

`increment?/5` now decides on three integers — guard digit, rest-nonzero (sticky folded in), and significand parity (`rem(signif, 2)` — the parity of an integer equals the parity of its last decimal digit) — instead of pattern-matching charlists. This is the classic guard/round/sticky formulation: a rounding decision never needs the digit sequence, only those three facts. The charlist helpers (`digits_increment`, `digits_to_integer`, `any_nonzero`) are deleted.

### 6. `round/3`: integer `do_round/5`

Same representation change as #5, sharing `split_digits/3` and the new `increment?/5`. Appending zeros becomes `coef * pow10(k)`. The old code needed two separate `exp < target_exp` branches because `:lists.split/2` requires an in-range index — when more digits were dropped than existed it first prepended literal zero characters. Integer `div`/`rem` has no such domain restriction (`div(coef, pow10(drop))` is simply 0 when `drop` exceeds the digit count), so that branch disappears.

The `normalize/1` call at the top of `round/3` is deliberately **kept**: removing it would change behavior for coefficients exceeding context precision (normalize applies context rounding before the placewise round), and with the new fast path its cost is negligible for in-precision values.

### 7. `sqrt/1`: digit count without list conversion

`integer_to_list` + `length` replaced by `coef_length/1`. A bit-length-based Newton seed was prototyped and **abandoned**: benchmarking showed the existing `shift` construction already scales the operand so the root lies within one decade of the fixed `pow10(precision + 1)` seed, so the "better" seed saved no iterations and cost an `encode_unsigned` per call. The seed logic is unchanged; its comment now explains why the fixed power of ten is a valid (and required) over-estimate. `sqrt`'s speedup comes from #3 and #5.

### 8. Context threading

`add`, `div`, and `sqrt` read the process-dictionary context two or three times per operation (e.g. `add` in `add_bounded?`, `add_bounded`, `add_sign`, and `context/3`). Each now reads it once at entry and passes the struct down via a new `context/4`. Not observable — the reads are within one synchronous call, so nothing can mutate the process dictionary between them — but it removes redundant lookups and struct copies. `Decimal.Context`'s public API is untouched.

---

## Compatibility

- **No public function changed** signature, spec, docs, or behavior. Everything deleted or renamed was private.
- **Signal flags are identical**, including the two division quirks above; flags are now asserted directly in tests for the first time in this suite.
- **Result representations are identical**, including trailing-zero preservation from exact division and the carry re-round from #238.
- Error messages are byte-identical.

## Testing

- Fixed two pre-existing assertion-less test bodies (`Context.with(... :floor)` blocks in the `add/2` and `sub/2` tests had bare `==` with no `assert`) — they were the only coverage of negative-zero under `:floor` rounding and have never tested anything.
- New unit tests: exact-quotient preferred exponents; the exact-but-wide quotient quirk (with flag assertions); `div_int`/`rem` quotient size limit including the exact `10^precision` boundary; the full rounding-mode table from the `Decimal.Context` moduledoc (10 values x 7 modes — previously verified nowhere); `round/3` when every digit is dropped, for the modes that lacked it; carry-at-context-precision (`99999 + 0.5` at precision 5); `:up` rounding at context precision (exact vs. inexact vs. sticky); `mult` precision rounding with flag assertions; `coef_length` boundaries (18/19 digits and 34/35 digits); `sqrt` across magnitudes 1e-100..1e101; `:xsd` rendering past the `pow10` table.
- Property tests: default generator coefficients widened from 10^16 to the full 34-digit decimal128 maximum, so every property now crosses the digit-count fallback and context-rounding paths. Four new properties: `div_int`/`rem` against `Kernel.div`/`Kernel.rem`; `div_rem` reconstruction (`q*b + r == a`, domains sized to keep it exact); floor/ceiling bracketing of all rounding modes; and `sqrt` of exact squares.
- The full suite additionally passed three consecutive runs with all properties at 1,000 iterations over the widened domain during development (committed run counts are the standard 100).

## Notes for reviewers

The highest-risk areas are `div_calc/4` (the closed-form equivalence to the old loop, including exact-quotient zero stripping) and `increment?/5` (the charlist-to-integer translation of each rounding mode). The rounding table test and the bracketing property were added specifically to make errors there loud.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Reviewed and prompted by me.

### DIFF
--- a/bench.exs
+++ b/bench.exs
@@ -23,16 +23,76 @@ decimals =
       do: struct(Decimal, %{sign: sign, coef: coef, exp: exp})
 
 decimal_pairs = for first <- decimals, second <- decimals, do: {first, second}
-binary_operations = %{"compare" => &Decimal.compare/2}
 
-jobs =
-  Map.new(binary_operations, fn {name, fun} ->
-    {name, fn -> Enum.each(decimal_pairs, fn {first, second} -> fun.(first, second) end) end}
-  end)
+# The full pair matrix (~10.7M pairs) is kept for `compare` so results stay
+# comparable with previously saved benchmarks. For the more expensive
+# arithmetic operations a deterministic sample keeps each benchee iteration
+# short enough to gather useful statistics.
+sampled_pairs = Enum.take_every(decimal_pairs, 199)
+
+positive_decimals = Enum.filter(decimals, &(&1.sign == 1))
+
+# High-precision operands at the decimal128 default precision (34 digits),
+# where division and rounding costs dominate.
+high_precision_coefs = [
+  1_234_567_890_123_456_789_012_345_678_901_234,
+  4_321_098_765_432_109_876_543_210_987_654_321,
+  9_999_999_999_999_999_999_999_999_999_999_999,
+  1_000_000_000_000_000_000_000_000_000_000_000
+]
+
+high_precision_decimals =
+  for coef <- high_precision_coefs,
+      exp <- [-40, -6, 0, 6, 40],
+      do: struct(Decimal, %{sign: 1, coef: coef, exp: exp})
+
+high_precision_pairs =
+  for first <- high_precision_decimals,
+      second <- high_precision_decimals,
+      do: {first, second}
+
+# Pairs with equal exponents and equal-length coefficients never short-circuit
+# on the adjusted exponent, so every comparison reaches the coefficient
+# alignment. This is the typical shape of same-scale comparisons, e.g. money
+# amounts at a fixed scale.
+same_scale_base = 1_234_567_890_123_456
+
+same_scale_decimals =
+  Enum.map(0..99, &struct(Decimal, %{sign: 1, coef: same_scale_base + &1, exp: -2}))
+
+same_scale_pairs =
+  for first <- same_scale_decimals,
+      second <- same_scale_decimals,
+      do: {first, second}
+
+each_pair = fn pairs, fun ->
+  fn -> Enum.each(pairs, fn {first, second} -> fun.(first, second) end) end
+end
+
+each = fn decimals, fun ->
+  fn -> Enum.each(decimals, fun) end
+end
+
+jobs = %{
+  "compare" => each_pair.(decimal_pairs, &Decimal.compare/2),
+  "compare same scale" => each_pair.(same_scale_pairs, &Decimal.compare/2),
+  "add" => each_pair.(sampled_pairs, &Decimal.add/2),
+  "sub" => each_pair.(sampled_pairs, &Decimal.sub/2),
+  "mult" => each_pair.(sampled_pairs, &Decimal.mult/2),
+  "div" => each_pair.(sampled_pairs, &Decimal.div/2),
+  "add high precision" => each_pair.(high_precision_pairs, &Decimal.add/2),
+  "mult high precision" => each_pair.(high_precision_pairs, &Decimal.mult/2),
+  "div high precision" => each_pair.(high_precision_pairs, &Decimal.div/2),
+  "round" => each.(decimals, &Decimal.round(&1, 2, :half_even)),
+  "normalize" => each.(decimals, &Decimal.normalize/1),
+  "sqrt" => each.(positive_decimals, &Decimal.sqrt/1),
+  "to_string scientific" => each.(decimals, &Decimal.to_string(&1, :scientific)),
+  "to_string normal" => each.(decimals, &Decimal.to_string(&1, :normal))
+}
 
 Benchee.run(jobs,
-  time: 20,
-  memory_time: 5,
+  time: 10,
+  memory_time: 2,
   save: [path: "benchmarks/#{tag}.benchee", tag: tag],
   formatters: [Benchee.Formatters.Console]
 )
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -363,27 +363,28 @@ defmodule Decimal do
   def add(%Decimal{} = num1, %Decimal{} = num2) do
     %Decimal{sign: sign1, coef: coef1, exp: exp1} = num1
     %Decimal{sign: sign2, coef: coef2, exp: exp2} = num2
+    ctx = Context.get()
 
     cond do
       coef1 == 0 and coef2 == 0 ->
-        sign = add_sign(sign1, sign2, 0)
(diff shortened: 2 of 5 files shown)


ericmj:
Beautiful, thank you!

PR #241, opened 2026-08-20
Performance: cut context, digit-count and float-conversion overhead

ismaelga:
Follow-up to #240, on the paths profiling showed still dominated. No public function changes: results, struct representations, error messages and the context flags (including order) are identical — checked by dumping the whole public surface for both builds and diffing, **146,539 cases, byte-identical.**

## Benchmark

`bench.exs`, baseline saved and loaded into the candidate run so both come from the same harness on the same machine (Apple M4 Max, Elixir 1.19.5 / OTP 28). Each row is one Benchee job over its whole workload.

| Operation | Before | After | Speedup |
| --- | --- | --- | --- |
| to_float | 126.7 μs | 4.73 μs | **26.8x** |
| mult | 15.9 ms | 6.1 ms | **2.6x** |
| add high precision | 511.5 μs | 209.5 μs | **2.4x** |
| money div | 137.8 μs | 58.5 μs | **2.4x** |
| add | 26.0 ms | 11.8 ms | **2.2x** |
| compare same scale | 452.7 μs | 216.3 μs | **2.1x** |
| div high precision | 391.4 μs | 194.0 μs | **2.0x** |
| mult high precision | 252.1 μs | 126.4 μs | **2.0x** |
| round | 988.5 μs | 511.5 μs | **1.9x** |
| sub | 26.0 ms | 14.0 ms | **1.9x** |
| div | 28.7 ms | 16.4 ms | **1.8x** |
| money add | 19.8 μs | 11.6 μs | **1.7x** |
| money compare | 7.98 μs | 5.06 μs | **1.6x** |
| money mult | 16.1 μs | 10.5 μs | **1.5x** |
| money round | 28.5 μs | 18.7 μs | **1.5x** |
| normalize | 287.7 μs | 201.5 μs | **1.4x** |
| compare | 614.8 ms | 458.0 ms | **1.3x** |
| from_float | 45.0 μs | 35.4 μs | **1.3x** |
| new from string | 20.9 μs | 16.4 μs | **1.3x** |
| sqrt | 2.2 ms | 1.7 ms | **1.3x** |
| to_string scientific | 501.9 μs | 439.8 μs | **1.1x** |
| inspect | 53.1 μs | 49.1 μs | **1.1x** |
| to_string normal | 510.3 μs | 474.2 μs | **1.1x** |

**Allocation per operation** (heap words; 1 word = 8 bytes), from `bench_resources.exs` — the deterministic per-operation figure, independent of how many iterations a job runs, unlike the workload-time table above.

| Operation | alloc before | after | reduction |
| --- | --- | --- | --- |
| to_float tiny | 279.6 w | 279.6 w | — |
| from_float | 200.6 w | 130.4 w | **1.54x** |
| sqrt wide | 125.5 w | 111.7 w | **1.12x** |
| div wide | 104.1 w | 82.2 w | **1.27x** |
| new money string | 95.7 w | 66.0 w | **1.45x** |
| add wide | 87.2 w | 68.9 w | **1.27x** |
| div money | 76.8 w | 55.3 w | **1.39x** |
| mult wide | 71.4 w | 57.1 w | **1.25x** |
| to_float money | 54.1 w | 49.2 w | **1.10x** |
| round money 2 | 51.9 w | 35.8 w | **1.45x** |
| sub money | 40.0 w | 32.4 w | **1.23x** |
| add money | 33.0 w | 25.4 w | **1.30x** |
| mult money | 30.2 w | 25.1 w | **1.20x** |
| normalize money | 27.1 w | 15.7 w | **1.72x** |
| to_string money | 20.1 w | 20.1 w | — |
| compare money | 7.4 w | 5.3 w | **1.39x** |

`to_float tiny` and `to_string` are unchanged in allocation: the first is dominated by building the `10^300` denominator (inherent to the conversion), and the second's output binary is the irreducible cost. The biggest spender, `from_float`, is the charlist round-trip through `:io_lib_format.fwrite_g/1`; a binary path exists but trades ~10% wall time for the memory, so it was left alone.

Reductions — the deterministic work count — drop 20-55%. `bench_resources.exs` reports wall, CPU, reductions, allocation, collector copying and peak footprint per operation, and is what caught two candidates that lowered allocation but cost wall time (rejected, below).

## What changed

- **Operations that signal nothing no longer touch the context.** The flag bookkeeping (wrap, fold, copy, process-dictionary write, trap scan) ran unconditionally — ~27% of `add/2`, ~50% of `round/3` — to conclude nothing happened. An empty signal list returns directly; an already-recorded signal skips the write.
- **The result's digit count is computed once per operation.** `precision/5` returns it; `exponent_limits/3` takes it; `add/2` no longer counts both operands' digits just to decide on the bounded path; `div/2` passes the count its own invariant pins to `precision + 1`.
- **`coef_length/1` compares against machine-word literals only.** Bignum comparisons cost an order of magnitude more, so walking 18 rungs before the bit-length estimate cost more than the estimate. 34-digit coefficients count ~3x faster; nothing counts slower.
- **Equal exponents short-circuit `compare/2` and `add/2`.** With equal exponents the coefficients decide alone; for `add/2` there is nothing to align, so the bounded path (the CVE-2026-32686 mitigation) has nothing to guard against — every exponent-gap case still does. A test pins that a 5,000-digit coefficient at an equal exponent stays bounded.
- **`to_float/1` scales with a computed shift.** Both scaling loops moved one bit per iteration, allocating a bignum each time (~50 for `1.5`, over a thousand near the range ends). The shift is the difference of the bit lengths, exact to within one bit.
- **Parsing reads the coefficient out of the scan.** Digits are accumulated straight into an integer while they fit a machine word, with no intermediate list. The exponent is bounded from its digit count, not by building a character list of the bound. Limits are still checked on counts before anything is converted — the property the CVE mitigation and #232 rely on.
- **Smaller ones:** `from_float/1` matches the redundant `.0` forward; dropping a single guard digit uses `div`/`rem`; `to_string/2` reuses its digit count; `Inspect` builds one binary; `normalize/1` returns its input struct when the coefficient has no trailing zero instead of rebuilding it (−42% allocation on `normalize`, and on `round` which calls it). The unreachable `base10?/1` branch in `div_calc/4` was removed (a nonzero remainder always signals `:inexact` through the sticky bit, so it could never change the flags).

## Compatibility

No public function changed signature, spec, docs, or behaviour; everything touched was private. Elixir 1.12.3/OTP 24 and 1.19.5/OTP 28 verified, `--warnings-as-errors` clean; nothing newer than Elixir 1.12 is used.

## Testing

The 146,539-case differential covers pairwise arithmetic over a coefficient/exponent matrix, all 7 rounding modes × 7 place counts, non-default precision/`emax`/`emin`/traps, 16,800 fuzzed parse inputs, 29,622 float conversions across the full exponent range, and multi-operation flag sequences. New unit tests and properties pin the changed paths against independent oracles (`compare` vs scaled-integer comparison, `to_float` vs the runtime's own conversion, `parse` vs building the coefficient from its digits); all pass against the previous implementation too.

## Prior work

Builds directly on work in the tree; several changes are re-tunings of code someone else introduced:

- **#240 (@tomciopp)** — `div_calc/4`, `split_digits/3`, `increment?/5`, square-and-multiply `pow10/1`, the bench script. Its note that the fixed `sqrt/1` seed already lands within a decade of the root checked out — a bit-length seed measured no faster, so `sqrt/1` is untouched.
- **#236 / #237 (@bariserdem)** — the sticky-bit division change, which is what makes the `base10?/1` branch above unreachable.
- **CVE-2026-32686 mitigation (97a23e6)** — `coef_length/1`, the bit-length digit-count estimate, the bounded addition path, chunked normalize. Three load-bearing properties are preserved: bignum digit counting stays estimate-based, parsing rejects on counts before converting, and every addition with an exponent gap still goes through the bounded path.
- **#232** — parse rejects over-long input before materializing; the rewritten scanner keeps that.
- **`to_float/1`'s algorithm (26e1443, 2016)**, including the Rick Regan reference — only the scaling loops changed.
- **Benchee**'s GC-trace allocation technique is reused by `bench_resources.exs`.

## Notes for reviewers

Highest-risk areas are `scale_up/3` / `scale_down/3` (the closed-form shift and its one-bit correction) and the digit count threaded through `precision/5` → `exponent_limits/3`; the `to_float` oracle test and the differential dump make errors there loud.

`bench_resources.exs` and `verify_diff.exs` are included as the measurement and verification tools; happy to drop either if they don't belong in the repo.

## Rejected candidates

- **`binary_to_integer/1` on input slices for the parser** — −32% reductions but +37% wall (BIF work is cheap in reductions, not in cycles). Accumulating during the scan won.
- **`:binary.match/2` to splice the redundant `.0` in `from_float/1`** — −39% reductions, 3.1x slower (pattern compiled per call).
- **Power-of-two Newton seed for `sqrt/1`** — no change, confirming #240's note.
- **Binary construction instead of iodata in `to_string/2`** — 1.5x slower (each nested construction copies what it has built).

---

🤖 Generated with [OpenCode](https://opencode.ai) using Claude. Prompted and reviewed by me.


ericmj:
This is great! Thank you! 💜

PR #243, opened 2026-08-25
Require Elixir 1.15 or later

BobbieBarker:
## Summary

- raise Decimal's minimum supported Elixir version from 1.12 to 1.15
- move the lower CI pair from Elixir 1.12 / OTP 24 to Elixir 1.15 / OTP 26
- record the compatibility change under the unreleased changelog

## Why this matters

ExDoc v0.40 requires Elixir 1.15 and adds native Markdown documentation and `llms.txt` generation. Raising the runtime floor separately keeps that compatibility decision explicit instead of hiding it inside a documentation-tool update.

This also moves the project's tested baseline onto a substantially newer compiler and OTP generation while retaining the existing current-runtime CI pair.

## Compatibility

This intentionally drops support for Elixir 1.12 through 1.14. It does not change Decimal's source or public API. The lower CI job now verifies the proposed minimum directly.

## Verification

- `mix format --check-formatted mix.exs`
- `MIX_ENV=test mix deps.unlock --check-unused`
- `MIX_ENV=test mix compile --warnings-as-errors`
- `MIX_ENV=test mix test` — 266 passed (105 doctests, 36 properties, 125 tests)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,8 @@ jobs:
       matrix:
         include:
           - pair:
-              elixir: "1.12"
-              otp: "24"
+              elixir: "1.15"
+              otp: "26"
           - pair:
               elixir: "1.19"
               otp: "28"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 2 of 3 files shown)


ericmj:
Please don't use automated agents to contribute.

PR #246, opened 2026-08-25
Resolve Elixir 1.20 type checker warnings

tomciopp:
Two sources, no behavior change:

The non guard expansions of `is_decimal/1` are a case with a %Decimal{}
clause and a catch all. When the argument is statically known to be a
%Decimal{} the type checker proves the catch all unreachable and warns
at every such call site, including the doctest. Marking the quoted
expansion `generated: true` declares the catch all deliberate,
following the convention for macro generated code.

Three tests pass intentionally invalid input to assert the raise
(new/1 with an atom, to_integer/1 with a NaN coefficient, round/1 with
nil). Routing those calls through apply/3 keeps the argument opaque to
the type checker, which does not check dynamic dispatch, while
exercising exactly the same failure path.

### DIFF
--- a/lib/decimal.ex
+++ b/lib/decimal.ex
@@ -277,7 +277,7 @@ defmodule Decimal do
     defmacro is_decimal(term) do
       case __CALLER__.context do
         nil ->
-          quote do
+          quote generated: true do
             case unquote(term) do
               %Decimal{} -> true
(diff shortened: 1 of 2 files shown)


ericmj:
Thank you! 💜

